PR #19, opened 2018-01-31
Update platforms.txt with NOAA-20, MSG 4, GOES-16

howff:


davidh-ssec:
What is this file used for?

adybbroe:
@davidh-ssec 
Mappings between satellite catalogue numbers and corresponding platform names from OSCAR

adybbroe:
It will allow you to do:
```
orb = Orbital('NOAA-20')
```

For NOAA-20 you currently need to do:
`orb = Orbital('NOAA 20')`

PR #20, opened 2018-01-31
Update index.rst with Installation section

howff:
For the platforms.txt file


adybbroe:
@howff Could you just change this PR to merge to develop instead?
Then I shall merge and update!

PR #31, opened 2018-08-09
Support xarray in get_observer_look

### LINKED ISSUE #30: get_observer_look turns xarray.DataArray objects into dask.array objects

```python

import datetime
import pyorbital.orbital
import xarray

lat = lon = xarray.DataArray([[1,2],[3,4]], dims=("x", "y"))
satlat = satlon = xarray.DataArray([[0]*2]*2, dims=("x", "y"))
times = xarray.DataArray([[datetime.datetime.now()]*2]*2, dims=("x", "y"))
satelevs = xarray.DataArray([[1000]*2]*2, dims=("x", "y"))
groundelevs = xarray.DataArray([[0]*2]*2, dims=("x", "y"))
times = xarray.DataArray([[datetime.datetime.now()]*2]*2, dims=("x", "y"))
(az, elev) = pyorbital.orbital.get_observer_look(satlon, satlat, satelevs,
                    times, lat, lon, groundelevs)
print("azimuth", az)
print("elev", elev)
```
#### Problem description

The `az` azimuth output is of type `dask.array`.  The coordinates and any other metadata stored alongside the arrays in the `DataArray` are lost.  I am not familiar with `dask.array`.

This is a consequence of the [if hasattr(az, 'chunks')](https://github.com/pytroll/pyorbital/blob/master/pyorbital/orbital.py#L114) block.  The `else:`-block should work fine for `xarray.DataArray` objects, which also happen to have this attribute.

#### Expected Output

I would expect both `az` and `elev` to be of `xarray.DataArray` type, such that no information is needlessly lost.

```
azimuth dask.array<rad2deg, shape=(2, 2), dtype=float64, chunksize=(2, 2)>
elev <xarray.DataArray (x: 2, y: 2)>
array([[79.686688, 69.883718],
       [60.955333, 53.06458 ]])
Dimensions without coordinates: x, y
```

Python 3.7.0, `pyorbital 1.3.1`, from conda-forge.

gerritholl:
get_observer_look was erroneously turning dask.array into xarray dataarrays..
However, the standard 'else:'-block did not work for xarray dataarrays.
either.  Add a special case 'elif:'-block for xarray dataarrays.

 - [x] Closes #30 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Passes ``git diff origin/develop **/*py | flake8 --diff`` 
 - [ ] Fully documented

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 language: python
 python:
 - '2.7'
-- "3.5"
 - "3.6"
 install:
+- pip install dask[array] xarray
 - pip install .
 - pip install coveralls
 script: coverage run --source=pyorbital setup.py test
--- a/pyorbital/orbital.py
+++ b/pyorbital/orbital.py
@@ -111,11 +111,15 @@ def get_observer_look(sat_lon, sat_lat, sat_alt, utc_time, lon, lat, alt):
 
     az_ = np.arctan(-top_e / top_s)
 
-    if hasattr(az_, 'chunks'):
-        # dask array
+    if hasattr(az_, 'chunks') and not hasattr(az_, 'loc'):
+        # dask array, but not xarray
         import dask.array as da
         az_ = da.where(top_s > 0, az_ + np.pi, az_)
         az_ = da.where(az_ < 0, az_ + 2 * np.pi, az_)
+    elif hasattr(az_, 'loc'):
+        # xarray
+        az_.data[top_s > 0] += np.pi
+        az_.data[az_.data < 0] += 2 * np.pi
     else:
         az_[top_s > 0] += np.pi
         az_[az_ < 0] += 2 * np.pi
--- a/pyorbital/orbital.py
+++ b/pyorbital/orbital.py
@@ -29,8 +29,21 @@ import warnings
 from datetime import datetime, timedelta
 
 import numpy as np
-
 from pyorbital import astronomy, dt2np, tlefile
+
+try:
+    import dask.array as da
+    has_dask = True
+except ImportError:
+    da = None
+    has_dask = False
+
+try:
+    import xarray as xr
+    has_xarray = True
+except ImportError:
+    xr = None
+    has_xarray = False
 
 ECC_EPS = 1.0e-6  # Too low for computing further drops.
 ECC_LIMIT_LOW = -1.0e-3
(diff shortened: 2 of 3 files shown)


djhoese:
Awesome, thanks for looking in to this. I'm wondering if we should just take the plunge and add some `isinstance` tests which we had originally tried to avoid due to how ugly the code might get. The dask functionality was added to have support for satpy's use of xarray.DataArray. I'm not sure if you are aware but xarray objects support dask arrays which is why this was done this way. A simpler solution for you in the short term may be to just pass `your_data_arr.data` to the `get_observer_look` function.

Really this function should check if we have a DataArray and if we do check if we are using dask underneath. If so, use the dask operations. If not, use the normal numpy operations. Although I think with recent versions of dask both can be treated in the same way as long as the syntax is supported.

djhoese:
@gerritholl What if you flipped the if statement and checked if it didn't have chunks and then did the normal numpy operations. How does xarray handle the numpy logic?:

```python
az_[top_s > 0] += np.pi
az_[az_ < 0] += 2 * np.pi
```

gerritholl:
@djhoese xarray doesn't handle the numpy logic in the multidimensional case.  I tried that first, but:

```
Traceback (most recent call last):
  File "mwe.py", line 12, in <module>
    times, lat, lon, groundelevs)
  File "/group_workspaces/cems2/fiduceo/Users/gholl/anaconda3/envs/FCDR37a/lib/python3.7/site-packages/pyorbital/orbital.py", line 120, in get_observer_look
    az_[top_s > 0] += np.pi
  File "/group_workspaces/cems2/fiduceo/Users/gholl/anaconda3/envs/FCDR37a/lib/python3.7/site-packages/xarray/core/dataarray.py", line 474, in __getitem__
    return self.isel(indexers=self._item_key_to_dict(key))
  File "/group_workspaces/cems2/fiduceo/Users/gholl/anaconda3/envs/FCDR37a/lib/python3.7/site-packages/xarray/core/dataarray.py", line 756, in isel
    ds = self._to_temp_dataset().isel(drop=drop, indexers=indexers)
  File "/group_workspaces/cems2/fiduceo/Users/gholl/anaconda3/envs/FCDR37a/lib/python3.7/site-packages/xarray/core/dataset.py", line 1427, in isel
    new_var = var.isel(indexers=var_indexers)
  File "/group_workspaces/cems2/fiduceo/Users/gholl/anaconda3/envs/FCDR37a/lib/python3.7/site-packages/xarray/core/variable.py", line 854, in isel
    return self[tuple(key)]
  File "/group_workspaces/cems2/fiduceo/Users/gholl/anaconda3/envs/FCDR37a/lib/python3.7/site-packages/xarray/core/variable.py", line 620, in __getitem__
    dims, indexer, new_order = self._broadcast_indexes(key)
  File "/group_workspaces/cems2/fiduceo/Users/gholl/anaconda3/envs/FCDR37a/lib/python3.7/site-packages/xarray/core/variable.py", line 473, in _broadcast_indexes
    self._validate_indexers(key)
  File "/group_workspaces/cems2/fiduceo/Users/gholl/anaconda3/envs/FCDR37a/lib/python3.7/site-packages/xarray/core/variable.py", line 520, in _validate_indexers
    "not supported. ".format(k.ndim))
IndexError: 2-dimensional boolean indexing is not supported.
```

gerritholl:
I assumed the reason for avoiding a bunch of `isinstance` statements was to prevent needlessly adding dependencies.  I'm aware xarray supports dask arrays, I haven't (explicitly) used that yet though.  

I know I could pass `.data` into the `get_observer_look` function, then reconstruct the relevant dimensions and coordinates afterwards.  I do that in some other places in my code, but I figured I might as well try to prevent the problem in the first place.

djhoese:
Good point I didn't think about the coordinates and that this function is producing new arrays, not adjusting existing ones. I'll try to find time to look at this tomorrow and add some tests. Thanks again for pointing this out. We technically don't support xarray objects in pyorbital, we just added a small hack for handling dask arrays better. May be something to consider in the future, but I'm not sure how much work it would be to make sure all functions handle xarray nicely.

djhoese:
@gerritholl I bit the bullet and added a `has_xarray` and `has_dask` variable based on import. Could you try this out and see if it works for you? I'm not sure I would argue that the code is cleaner, but it *may* be easier to understand now.

djhoese:
FYI This PR was accidentally closed when the `develop` branch was removed (we went to a master-only branch scheme). I have merged this branches changes in to master now but github doesn't seem like it recognizes that.

So, this has been merged not closed/unmerged.

PR #49, opened 2019-10-18
Equatorial Crossing Time

sfinkens:
Add method to compute the equatorial crossing time of an orbit. 

A utility to read pygac-like TLE files (TLEs concatenated over time) can be added later.

 - [x] Tests added 
 - [x] Tests passed 
 - [x] Fully documented


sfinkens:
@mraspaud Of course! I'm also thinking about a switch to compute the crossing time of the descending node.

adybbroe:
@sfinkens Should we add some tests and get this merged soon? I may be able to find some time to help out. As I am in closing/merging mode atm :-)

sfinkens:
@adybbroe Instead of using the orbit number I tried to find the root of the cartesian z-coordinate (with constraint z-velocity > 0 for ascending node and < 0 for descending node). But I wasn't able to convince the optimization algorithm to converge...

adybbroe:
Great, thanks @sfinkens This will be a useful feature. We will see if the orbit-number instability as I mentioned will become an issue

PR #56, opened 2020-01-29
Add a script to download TLEs and store them to a database

pnuu:
This PR adds a script that can be used to download TLE data from various sources, and store them in a database. The latest TLEs can also stored to a text file.

**Testing for Python 2.7 is removed!**

 - [x] Tests added 
 - [x] Tests passed 
 - [x] Passes ``flake8 pyorbital`` for the affected parts
 - [x] Fully documented


djhoese:
This makes me think of the `<prefix>/share/proj/proj.db` database that the PROJ C library installs. However, PROJ doesn't write to the database once it is installed as far as I know but I don't use the C library very much so there may be an API for that.

Maybe this database should be stored in a location like the PROJ database, but the TLE files are stored in a more temporary or cache-y location? Just some ideas.

pnuu:
I think it's best to let the user decide where the database is stored. All the software we have (pyorbital, AAPP, PPS, pytroll schedule) read the TLE data from text files and either use `TLES` environment variable or their own directory structure and expect the files to be there, without a possibility to configure it.

I'll sill adjust the text file saving so that multiple output directories/patterns are possible.

djhoese:
> or their own directory structure and expect the files to be there

Could there be an API in pyorbital or other package that given parameters (platform, time, etc) gives back the TLE file or the information from the TLE file?

PR #84, opened 2021-10-15
add get all platforms function and generator getting one platform to processing

19as:
- [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
 Pyorbital
 =========
 
-[![Build status](https://github.com/pytroll/pyorbital/workflows/CI/badge.svg?branch=master)](https://github.com/pytroll/pyorbital/workflows/CI/badge.svg?branch=master)
-[![Coverage Status](https://coveralls.io/repos/github/pytroll/pyorbital/badge.svg?branch=master)](https://coveralls.io/github/pytroll/pyorbital?branch=master)
+[![Build status](https://github.com/pytroll/pyorbital/workflows/CI/badge.svg?branch=main)](https://github.com/pytroll/pyorbital/workflows/CI/badge.svg?branch=main)
+[![Coverage Status](https://coveralls.io/repos/github/pytroll/pyorbital/badge.svg?branch=main)](https://coveralls.io/github/pytroll/pyorbital?branch=main)
 [![PyPI version](https://badge.fury.io/py/pyorbital.svg)](https://badge.fury.io/py/pyorbital)
 
 
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,6 +1,6 @@
 # Releasing Pyorbital
 
-1. checkout master
+1. checkout main branch
 2. pull from repo
 3. run the unittests
 4. run `loghub` and update the `CHANGELOG.md` file:
--- a/pyorbital/etc/platforms.txt
+++ b/pyorbital/etc/platforms.txt
@@ -69,6 +69,7 @@ RadarSat-2 32382
 Sentinel-1A 39634
 Sentinel-3A 41335
 Sentinel-3B 43437
+Sentinel-5P 42969
 SMOS 36036
 SPOT-5 27421
 SPOT-6 38755
--- a/pyorbital/orbital.py
+++ b/pyorbital/orbital.py
@@ -126,27 +126,17 @@ def get_observer_look(sat_lon, sat_lat, sat_alt, utc_time, lon, lat, alt):
     top_z = cos_lat * cos_theta * rx + \
         cos_lat * sin_theta * ry + sin_lat * rz
 
-    az_ = np.arctan(-top_e / top_s)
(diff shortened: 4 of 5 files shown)


19as:
please, merge my request in trunk

.gitignore
```diff
@@ -39,3 +39,4 @@ nosetests.xml
 
 # rope
 .ropeproject
+/.vs
```
djhoese, line 42:
```suggestion
.vs
```

I don't think the `/` is needed.

pyorbital/tlefile.py
```diff
@@ -106,6 +106,44 @@ def read(platform, tle_file=None, line1=None, line2=None):
     return Tle(platform, tle_file=tle_file, line1=line1, line2=line2)
 
 
+def get_all_platforms(tle_file=None):
+    """Read all *platform* name from tle file
+    tle_file - list path files
+    """
```
djhoese, line 112:
```suggestion
    """Read all *platform* names from tle file."""
```

pyorbital/tlefile.py
```diff
@@ -106,6 +106,44 @@ def read(platform, tle_file=None, line1=None, line2=None):
...
+            l_2 = next(fid).decode('utf-8')
+    fid.close()
+
+    return platforms
+
+
+def get_platform(tle_file=None):
+    """Read *platform* name from tle file
+    tle_file - path to tle file
+    """
```
djhoese, line 130:
```suggestion
    """Read *platform* name from tle file."""
```

pyorbital/tlefile.py
```diff
@@ -106,6 +106,44 @@ def read(platform, tle_file=None, line1=None, line2=None):
...
+    try:
+        fid = io.open(tle_file, 'rb')
+        for line in fid:
+            line = line.decode('utf-8')
+            line = line.strip()
+            if not line.startswith('1') and not line.startswith('2') and len(line) > 0:
+                yield line
+                l_1 = next(fid).decode('utf-8')
+                l_2 = next(fid).decode('utf-8')
```
djhoese, line 139: Why decode these lines and assign them to a variable if they aren't used?

pyorbital/tlefile.py
```diff
@@ -106,6 +106,44 @@ def read(platform, tle_file=None, line1=None, line2=None):
...
+
+    return platforms
+
+
+def get_platform(tle_file=None):
+    """Read *platform* name from tle file
+    tle_file - path to tle file
+    """
+    try:
+        fid = io.open(tle_file, 'rb')
```
djhoese, line 132: What about getting rid of the `try/except/finally` and instead using `with io.open(tle_file, 'rb') as fid:`?

djhoese (changes requested):
Sorry about the delay, we've been kind of busy and sometimes pull requests get lost in the inbox. How do you plan on using this functionality?

I know very little about TLE files so correct me if I'm missing something, but it looks like your two functions are very similar and possibly don't do what they say. The top one says that it gets multiple platforms and the bottom one says it gets a single one. However, the second one is a generator so it actually yields all the platforms too.

Lastly, these functions are almost exactly the same code. Could they be combined? Or one could use the other? If the only difference between these is a list versus a generator then I would suggest having only the generator and leave it up to the user to call `list(get_platform(tle_file))` instead of doing it for them.

djhoese:
Oh also could you try adding some tests for these functions?

djhoese:
@19as Any updates on this? I had some questions and comments that I was hoping you could answer before we move forward with this pull request. If you have questions for me let me know. If you don't reply in the next week or two then I will plan on closing this.

19as:
> Sorry about the delay, we've been kind of busy and sometimes pull requests get lost in the inbox. How do you plan on using this functionality?
> 
> I know very little about TLE files so correct me if I'm missing something, but it looks like your two functions are very similar and possibly don't do what they say. The top one says that it gets multiple platforms and the bottom one says it gets a single one. However, the second one is a generator so it actually yields all the platforms too.
> 
> Lastly, these functions are almost exactly the same code. Could they be combined? Or one could use the other? If the only difference between these is a list versus a generator then I would suggest having only the generator and leave it up to the user to call `list(get_platform(tle_file))` instead of doing it for them.


You're right. It is better to use one generator. I had to go through the file and, upon receiving one aircraft, process it

pyorbital/tlefile.py
19as: Sorry, It is used to debug

pyorbital/tlefile.py
19as: Of course, it is really worked

19as:
> Oh also could you try adding some tests for these functions?

я постараюсь это сделать в будущем

PR #102, opened 2022-07-07
Drop direct support of Python <3.8

djhoese:
We don't have any direct incompatibilities currently, but we need to define this sooner rather than later so when we do users are able to download incompatible versions from PyPI.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented


mraspaud (approved):
LGTM

PR #103, opened 2022-07-11
www.celestrak.org → celestrak.org

gerritholl:
After the http→https change in #101, the www. should be dropped as well
to avoid SSL certificate errors/warnings.  Change www.celestrak.org to
celestrak.org for those cases.

See also the post-merge comments on #101.

 - [x] Closes #101 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented


mraspaud (approved):
LGTM

PR #128, opened 2023-05-30
Fix typo in VIIRS geoloc definition

frdcms:
This PR fixes a typo in the code for VIIRS geoloc definition.

### DIFF
--- a/pyorbital/tests/test_geoloc.py
+++ b/pyorbital/tests/test_geoloc.py
@@ -190,6 +190,16 @@ class TestGeolocDefs:
         np.testing.assert_allclose(geom.fovs,
                                    expected_fovs, rtol=1e-2, atol=1e-2)
 
+    def test_viirs_defaults(self):
+        """Test the definition of the viirs instrument with default slicing."""
+        geom = viirs(1, chn_pixels=3)
+        expected_fovs = np.array([
+            np.tile(np.array([[0.98, -0., -0.98]]), [32, 1]),
+            np.tile(np.array([[0., -0., 0]]), [32, 1])], dtype=np.float64)
+
+        np.testing.assert_allclose(geom.fovs,
+                                   expected_fovs, rtol=1e-2, atol=1e-2)
+
     def test_amsua(self):
         """Test the definition of the amsua instrument."""
         geom = amsua(1)
--- a/pyorbital/geoloc_instrument_definitions.py
+++ b/pyorbital/geoloc_instrument_definitions.py
@@ -156,7 +156,7 @@ def viirs(scans_nb, scan_indices=slice(0, None),
     """
 
     entire_width = np.arange(chn_pixels)
-    scan_points = entire_width[scan_indices.astype('int')]
+    scan_points = entire_width[scan_indices].astype('int')
     scan_pixels = len(scan_points)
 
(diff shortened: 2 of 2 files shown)


djhoese:
I'm unfamiliar with this part of the code, but could someone tell me where/when this function is used? How was this not caught before? `scan_indices` defaults to a `slice` object so this never should have worked in the default case. And if we assume `chn_pixels` is an integer then the `astype` shouldn't be needed anyway.

Thanks @frdcms for bringing this to our attention.

adybbroe (approved):
LGTM! Thanks for spotting it!
Are there no tests for this currently, seems not?

adybbroe:
I agree with @djhoese that this is a bit strange it has been sitting there for so long. I do not recall it. And I have mostly been working with the sounder instruments over the last years. And apparently we do not use this function in our real-time production, otherwise we should have seen it. So go ahead, please.

frdcms:
Thanks @adybbroe 
The tests in `pytorbital.tests.test_geoloc.TestGeolocDefs.test_viirs` all gave the argument `scan_indices` to `geoloc_instrument_definitions.viirs` so the bug was not highlighted.
Just added another test in order to use the default value for `scan_indices` in `geoloc_instrument_definitions.viirs`.

pyorbital/tests/test_geoloc.py
```diff
@@ -190,6 +190,16 @@ def test_viirs(self):
...
+        geom = viirs(1, chn_pixels=3)
+        expected_fovs = np.array([
+            np.tile(np.array([[0.98, -0., -0.98]]), [32, 1]),
+            np.tile(np.array([[0., -0., 0]]), [32, 1])], dtype=np.float64)
+
+        np.testing.assert_allclose(geom.fovs,
+                                   expected_fovs, rtol=1e-2, atol=1e-2)
+        #print(geom.fovs)
+        #print(geom.fovs.shape)
+
```
djhoese, line 202:
```suggestion
    def test_viirs_defaults(self):
        """Test the definition of the viirs instrument with default slicing."""
        geom = viirs(1, chn_pixels=3)
        expected_fovs = np.array([
            np.tile(np.array([[0.98, -0., -0.98]]), [32, 1]),
            np.tile(np.array([[0., -0., 0]]), [32, 1])], dtype=np.float64)

        np.testing.assert_allclose(geom.fovs,
                                   expected_fovs, rtol=1e-2, atol=1e-2)

```

djhoese: Our new-ish standard practice is to make separate tests actually separate rather than having a ton of independent test in a single test method.

frdcms: OK, suggested changes just pushed in a squashed commit.

frdcms:
Hello @adybbroe,
Do you have any idea when the next release will be created?

frdcms:
Sorry to insist but I need to know whether the next release will be created soon or later to deploy this fix in a production environment.
Thanks in advance for your answer.

djhoese:
This should be included in the 1.8.0 release that is on PyPI. The package will get to conda-forge later today hopefully.

frdcms:
Great!
Thank you very much.

PR #138, opened 2023-12-31
Update celestrak urls

dagnic:
Celestrak starts to remove the legacy tle txt files. I've detected a 404 error on the active one and asked T.S. Kelso; he points me to [this post](https://twitter.com/TSKelso/status/1740560827378483612) on X. I now suggest to use the query links. 

 - [ ] Closes https://github.com/pytroll/pyorbital/issues/139
 - [x] Tests passed 

### DIFF
--- a/pyorbital/tlefile.py
+++ b/pyorbital/tlefile.py
@@ -37,16 +37,18 @@ import sqlite3
 from xml.etree import ElementTree as ET
 from itertools import zip_longest
 
+TLE_GROUPS = ('active',
+              'weather',
+              'resource',
+              'cubesat',
+              'stations',
+              'sarsat',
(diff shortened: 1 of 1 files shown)


djhoese:
How urgent is this fix needed? I have almost zero experience with this package but I also don't remember when all of the people who do have experience are coming back from holiday time off.

dagnic:
I would say quite urgent because without the fix, it is not possible to get the satellite orbital parameters through the API. If you need time, one can either: 
- use my fix directly
- manually downloaded the TLE file from celestrak website using the new link and just load the file with the API.

adybbroe (approved):
LGTM. Many thanks for this!

PR #157, opened 2024-07-03
Update `get_next_passes` docstring

simonrp84:
Currently, the documentation for `get_next_passes` is unclear. It does not specify what units should be used for the `altitude`. This PR updates the docstring to make clear that `altitude` should be in km.

 - [x] Fully documented


pnuu (approved):
Thanks!

PR #158, opened 2024-07-10
Update astronomy.py

Yusufibin:
Error handling has been optimized with the observer_position function, which now checks the type and value of the altitude, and raises appropriate exceptions in case of invalidity. Important astronomical constants are defined at the beginning of the file to improve code readability and maintainability. Type annotations have been added for better understanding and readability of the code.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/pyorbital/astronomy.py
+++ b/pyorbital/astronomy.py
@@ -1,11 +1,12 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 #
-# Copyright (c) 2011, 2013
+# Copyright (c) 2011, 2013, 2023
 #
 # Author(s):
 #
 #   Martin Raspaud <martin.raspaud@smhi.se>
+#   Bardur Arantsson <bardur.arantsson@smhi.se>
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


pyorbital/astronomy.py
```diff
@@ -39,55 +40,97 @@
...
+EARTH_ flattening = 1 / 298.257223563  # Earth flattening WGS-84
+EARTH_RADIUS = 6378.137  # WGS84 Equatorial radius in km
 MFACTOR = 7.292115E-5
+EARTH_SEMI_MAJOR_AXIS = 1.00000261 * AU  # Earth semi-major axis in meters
+EARTH_ECCENTRICITY = 0.01671123  # Earth orbital eccentricity
+EARTH_YEAR = 365.25636  # Length of Earth year in days
+PERIHELION_DAY = 3  # Day of year with Earth in perihelion (approx.)
 
+# Type alias for time arguments
+TimeLike = Union[dt.datetime, np.datetime64, np.ndarray]
```
djhoese, line 58: There is the `numpy.typing` module added in numpy 1.20 I believe. Could this be used to define the dtype of the ndarray? Using something like `npt.ArrayLike[np.datetime64]`?

pyorbital/astronomy.py
```diff
@@ -39,55 +40,97 @@
...
 
-def jdays2000(utc_time):
+
+def jdays2000(utc_time: TimeLike) -> Union[float, np.ndarray]:
     """Get the days since year 2000.
+
+    :param utc_time: UTC time
+    :type utc_time: datetime.datetime, numpy.datetime64, or array-like
+    :return: Days since year 2000
+    :rtype: float or numpy.ndarray
```
djhoese, line 67: I'm not sure if pyorbital has a defined docstring format it follows, but in Satpy we've been using Google style docstrings:

https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

Would it be possible to reformat these to match? If other parts of pyorbital are using numpy style docstrings then I'd be open to that as well.

pyorbital/astronomy.py
```diff
@@ -39,55 +40,97 @@
...
     """
     return jdays2000(utc_time) + 2451545.0
 
 
-def _days(dt):
+def _days(dt: TimeLike) -> Union[float, np.ndarray]:
     """Get the days (floating point) from *d_t*.
+
+    :param dt: Time delta
+    :type dt: datetime.timedelta or numpy.timedelta64
```
djhoese, line 87: Some of these type annotations and docstrings are incorrect I think. This function for example takes `datetime.timedelta` objects or a numpy array of `timedelta64` objects. You have `TimeLike` listed here.

pyorbital/astronomy.py
djhoese: Additionally, I think you can remove the type in the docstrings if you have the annotations. We'll need to possibly enable the correct sphinx options to copy the annotations to the rendered documentation, but that can come after everything else is sorted.

pyorbital/astronomy.py
```diff
@@ -39,55 +40,97 @@
...
 import numpy as np
+from typing import Tuple, Union
 
 from pyorbital import dt2np
 
-F = 1 / 298.257223563  # Earth flattening WGS-84
-A = 6378.137  # WGS84 Equatorial radius
+# Astronomical constants
+AU = 149597870700.0  # Astronomical unit in meters
+EARTH_ flattening = 1 / 298.257223563  # Earth flattening WGS-84
```
djhoese, line 49: Typo here (extra space). Also why not uppercase for the flattening part?

djhoese (changes requested):
Nice attempt at tackling this. I had a couple suggestions and changes I'd like to see if we plan on merging this. Also note that I'm pretty in favor of annotations but others on the maintainer team may not enjoy all the clutter it adds. Especially for this module where so many types are possible as arguments.

PR #159, opened 2024-07-17
Bugfix use of datetime.datetime objects in call to `get_last_an_time`

adybbroe:
Fix so a normal datetime.datetime object can be passed as well to the get last ascending node function.

Currently when running this code below:

```
import pyorbital.orbital
from datetime import datetime

tlefile = "/path/to/some/archived/tle/files/tle-202406250015.txt"

dtobj = datetime(2024, 6, 25, 11, 0, 18)
orb = pyorbital.orbital.Orbital('NOAA-20', tle_file=tlefile)
result = orb.get_last_an_time(dtobj)
```

you will get this kind of error:

```
File <some path>/pyorbital/pyorbital/orbital.py:172, in Orbital.get_last_an_time(self, utc_time)
    170 dt = np.timedelta64(10, 'm')
    171 t_old = utc_time
--> 172 t_new = t_old - dt
    173 pos0, vel0 = self.get_position(t_old, normalize=False)
    174 pos1, vel1 = self.get_position(t_new, normalize=False)

UFuncTypeError: ufunc 'subtract' cannot use operands with types dtype('O') and dtype('<m8[m]')
```

THis PR seeks to solve these kind of errors, allowing to pass a simple `datetime.datetime` object to the function, and not only `numpy.datetime64` objects as seems to have been the case until now.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/pyorbital/orbital.py
+++ b/pyorbital/orbital.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
-# Copyright (c) 2011-2023 Pyorbital developers
+# Copyright (c) 2011-2024 Pyorbital developers
 
 # Author(s):
 
@@ -86,11 +86,14 @@ A3OVK2 = (-XJ3 / CK2) * AE**3
 
 
 class OrbitalError(Exception):
+    """Custom exception for the Orbital class."""
+
     pass
 
 
 def get_observer_look(sat_lon, sat_lat, sat_alt, utc_time, lon, lat, alt):
     """Calculate observers look angle to a satellite.
+
     http://celestrak.com/columns/v02n02/
 
     :utc_time: Observation time (datetime object)
@@ -142,7 +145,6 @@ def get_observer_look(sat_lon, sat_lat, sat_alt, utc_time, lon, lat, alt):
 
 
 class Orbital(object):
-
     """Class for orbital computations.
 
     The *satellite* parameter is the name of the satellite to work on and is
@@ -151,6 +153,7 @@ class Orbital(object):
     """
 
     def __init__(self, satellite, tle_file=None, line1=None, line2=None):
+        """Initialize the class."""
         satellite = satellite.upper()
         self.satellite_name = satellite
         self.tle = tlefile.read(satellite, tle_file=tle_file,
@@ -159,16 +162,14 @@ class Orbital(object):
         self._sgdp4 = _SGDP4(self.orbit_elements)
 
     def __str__(self):
+        """Print the Orbital object state."""
         return self.satellite_name + " " + str(self.tle)
 
     def get_last_an_time(self, utc_time):
-        """Calculate time of last ascending node relative to the
-        specified time
-        """
-
+        """Calculate time of last ascending node relative to the specified time."""
         # Propagate backwards to ascending node
         dt = np.timedelta64(10, 'm')
-        t_old = utc_time
+        t_old = np.datetime64(utc_time)
         t_new = t_old - dt
         pos0, vel0 = self.get_position(t_old, normalize=False)
         pos1, vel1 = self.get_position(t_new, normalize=False)
@@ -236,14 +237,17 @@ class Orbital(object):
         return np.rad2deg(lon), np.rad2deg(lat), alt
 
     def find_aos(self, utc_time, lon, lat):
+        """Find AOS."""
         pass
 
     def find_aol(self, utc_time, lon, lat):
+        """Find AOL."""
         pass
 
     def get_observer_look(self, utc_time, lon, lat, alt):
         """Calculate observers look angle to a satellite.
-        http://celestrak.com/columns/v02n02/
+
+        See http://celestrak.com/columns/v02n02/
 
         utc_time: Observation time (datetime object)
         lon: Longitude of observer position on ground in degrees east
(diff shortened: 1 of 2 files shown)


pyorbital/orbital.py
```diff
@@ -159,16 +162,14 @@ def __init__(self, satellite, tle_file=None, line1=None, line2=None):
...
     def get_last_an_time(self, utc_time):
-        """Calculate time of last ascending node relative to the
-        specified time
-        """
-
+        """Calculate time of last ascending node relative to the specified time."""
         # Propagate backwards to ascending node
         dt = np.timedelta64(10, 'm')
-        t_old = utc_time
+        t_old = np.datetime64(utc_time)
```
djhoese, line 172: FYI in modern python times a datetime object is "supposed to" have a timezone defined on it. If you use this method of conversion here and a python datetime object has a timezone then you'll get a warning:

```
In [13]: now = dt.datetime.now(dt.timezone.utc)

In [14]: np.datetime64(now)
<ipython-input-14-019101e2dae5>:1: UserWarning: no explicit representation of timezones available for np.datetime64
  np.datetime64(now)
Out[14]: np.datetime64('2024-07-17T15:36:14.860486')

```

I *think* you could check if the datetime has a `.tzinfo` and if not then convert it to UTC and then drop the timezone (not sure how to do that yet) before passing it to `np.datetime64` to avoid the warning. Or I guess you could drop any timezone information no matter what because you have defined this interface as expecting time in UTC. :thinking:

pyorbital/tests/test_orbital.py
```diff
@@ -407,3 +415,23 @@ def test_63(self):
...
         warnings.filterwarnings('default')
+
+
+@pytest.mark.parametrize('dtime, expected',
+                         [(datetime(2024, 6, 25, 11, 0, 18),
+                           np.datetime64('2024-06-25T10:44:18.234375')),
+                          (datetime(2024, 6, 25, 11, 5, 0),
+                           np.datetime64('2024-06-25T10:44:18.234375')),
+                          (np.datetime64('2024-06-25T11:10:00.000000'),
+                           np.datetime64('2024-06-25T10:44:18.234375')),
```
sfinkens, line 426: The expected value is always the same, so you could define that in the test method.

pyorbital/orbital.py
adybbroe: Good point, thanks! I believe I am handling this now!?

pyorbital/tests/test_orbital.py
adybbroe: Oh, yes, you are right, the truth is constant here. Fixed.

pyorbital/orbital.py
```diff
@@ -917,7 +921,26 @@ def propagate(self, utc_time):
...
+    The input *utc_time* is either a timezone unaware object assumed to be in
+    UTC, or a timezone aware datetime object in UTC.
+    """
+    if not hasattr(utc_time, 'tzinfo') or utc_time.tzinfo is None:
+        return utc_time
+
+    if utc_time.tzinfo != pytz.utc:
+        raise AttributeError("UTC time expected! Parsing a timezone aware datetime object requires it to be UTC!")
+
+    return utc_time.replace(tzinfo=None)
```
sfinkens, line 936: I think readability could be improved by separating the datetime path from the datetime64 path, for example 

```python
if isinstance(utc_time, dt.datetime):
    if utc_time.tzinfo and utc_time.tzinfo != pytz.utc:
        raise AttributeError
    return utc_time.replace(tzinfo=None)
return utc_time
```

sfinkens (commented):
Looking good! One minor comment regarding readability

pyorbital/orbital.py
```diff
@@ -917,7 +921,25 @@ def propagate(self, utc_time):
...
 
+def _get_tz_unaware_utctime(utc_time):
+    """Return timzone unaware datetime object.
+
+    The input *utc_time* is either a timezone unaware object assumed to be in
+    UTC, or a timezone aware datetime object in UTC.
+    """
+    if isinstance(utc_time, datetime):
+        if utc_time.tzinfo and utc_time.tzinfo != pytz.utc:
+            raise AttributeError("UTC time expected! Parsing a timezone aware datetime object requires it to be UTC!")
```
djhoese, line 932: I think this should be a `ValueError`. The `AttributeError` I think is more to say you tried to access an attribute that didn't exist. `ValueError` (in my opinion) is more like "you gave me a value that doesn't make sense".

pyorbital/orbital.py
```diff
@@ -917,7 +921,25 @@ def propagate(self, utc_time):
...
+        return utc_time.replace(tzinfo=None)
+
+    return utc_time
+
+
 def kep2xyz(kep):
+    """Keppler to cartesian coordinates conversion.
+
+    (Not sure what 'kep' acttually refers to, just guessing! FIXME!)
```
djhoese, line 941:
```suggestion
    (Not sure what 'kep' actually refers to, just guessing! FIXME!)
```

djhoese (changes requested):
See my other previous comments for requested changes.

sfinkens:
Agree with the value error, otherwise looks good to me!

PR #169, opened 2024-11-20
Switch to use pyproject.toml instead of setup.py, and skip versioneer

adybbroe:
This PR started out solving a deprecation warning when installing with pip:

```
  DEPRECATION: Legacy editable install of pyorbital==1.8.3+29.g19f444b from file:///home/a000680/usr/src/forks/pyorbital (setup.py develop) is deprecated. pip 25.0 will enforce this behaviour change. A possible replacement is to add a pyproject.toml or enable --use-pep517, and use setuptools >= 64. If the resulting installation is not behaving as expected, try using --config-settings editable_mode=compat. Please consult the setuptools documentation for more information. Discussion can be found at https://github.com/pypa/pip/issues/11457
```

It ended up switching to pyproject.toml, ditching versioneer.py and by using more strict style checkers with ruff many small issues needed attention.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented


djhoese:
As discussed the one remaining ruff error was a complexity issue with `pyorbital/orbital.py:_SGDP4.__init__`. I've added an ignore on that method to get ruff to pass. Dealing with the complexity of that method will take a considerable amount of work. My understanding is @pnuu might tackle it if he's foolish enough.

pnuu:
I'm actually already browsing through `_SGDP4.__init__`. If I have a moment tomorrow I'll have a go at it.

PR #172, opened 2024-11-29
Update CI build action to work without setup.py

pnuu:
In https://github.com/pytroll/pyorbital/pull/169 the `setup.py` was removed, but the CI action to build and deploy the released package was updated. This PR adapts the action to use `python -m build` command.


mraspaud (approved):
Thanks for the fix, LGTM!

PR #181, opened 2025-01-30
Add platforms to platforms.txt to support Argos network

seapyth0n:
While the platform list does support some of the Argos platforms, it doesn't support all. Three additional IDs were added:
EOS-6 (OCEANSAT-3) 54361  
SARAL 39086
OTB-3-GAZELLE 54023

ANGELS is also an Argos satellite but has been decommissioned. This platform is not currently in the platforms.txt and was not added. 

Information about the Argos platforms was collected here: https://celestrak.org/NORAD/elements/gp.php?GROUP=argos&FORMAT=tle


mraspaud (approved):
Thanks a lot for this and sorry for the late reply... Merging

PR #186, opened 2025-04-04
Support yaw arrays for ScanGeometry with same broadcasting as roll / pitch

oembury:
ScanGeometry.vectors has optional arguments for specifying roll, pitch and yaw rotations.

roll and pitch could already be specified as arrays which would be broadcast with the scan geometry using standard numpy broadcasting rules. However, yaw could only be passed as a scalar (as it was passed through unmodified to qrotate)

This PR adds a simple check to detect an array-like argument for yaw and ensure that it will follow the same broadcast rules already used for the roll and pitch arguments.

 - [x] Tests added


mraspaud (approved):
LGTM

PR #202, opened 2025-09-26
Move import of tlefile inline to avoid import-time side effects

djhoese:
Importing `pyorbital.orbital` was importing `tlefile` which was causing config file checking and loading even though much of the functionality in `orbital.py` doesn't use it.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented


mraspaud (approved):
LGTM

PR #206, opened 2025-10-03
Refactor orbital tracking with vectorized methods, bug fixes and improved utilities

### LINKED ISSUE #95: Would be nice to have a `get_last_dn_time` to get the last descending node

```python
# Your code here

```
#### Problem description

Would be nice to have a `get_last_dn_time` to get the last descending node similar to the existing `get_last_an_time` method.

JaskRendix:
PR improves the orbital tracking module and related utilities.

- fixed a bug in the main script where observer longitude and latitude were being converted to radians before being passed to `get_observer_look`, which already performs this conversion internally, coordinates are now passed in degrees as expected, and conversion is handled inside the method
- refactored the `get_observer_look` function: split the logic into two helper functions: `ecef_to_topocentric` and `compute_azimuth_elevation`, this isolates coordinate transformations and makes the azimuth/elevation computation easier to test and reuse, pytest coverage added
- refactored `_get_root` and `_get_max_parab`: `_get_root` now includes a check for valid root bracketing and uses `xtol` and `rtol` for better convergence control. `_get_max_parab` now includes a fallback for flat functions, a maximum iteration limit, and improved numerical stability to avoid divergence and handle edge cases.
- cleaned up `test_orbital`: ran `isort` and `black` for formatting and added new pytest coverage for all newly introduced vectorized methods and utilities
- implemented `find_aos` and `find_aol` methods for detecting Acquisition of Signal and Loss of Signal events based on observer location and elevation mask; includes robust horizon-crossing logic and full pytest coverage
- improved `get_last_an_time` logic by refactoring shared node detection into a private `_find_last_node_time` method, which uses direct geometric analysis of the satellite’s Z-position and velocity to identify node crossings; this approach is distinct from orbit-number-based equatorial crossing methods and supports both ascending and descending detection, added `get_last_dn_time` to retrieve the last descending node, addressing feature request #95; includes parameterized pytest coverage for velocity sign, node classification, and temporal accuracy

 - [x] Closes #95 
 - [x] Tests added
 - [x] Fully documented


djhoese (commented):
Nice. At a high level I have two big questions:

1. Is it possible to make the original methods work vectorized without needed separate "vectorized" specific methods?
2. Would using something like numpy.vectorize help? https://numpy.org/doc/stable/reference/generated/numpy.vectorize.html. I think in this case the outputs would always be numpy arrays but I think that's OK.

One other concern is the use of `np.array` to convert to array types and `isinstance` checks on numpy arrays. It'd be nice if these functions were compatible with dask arrays which likely (hopefully) just treating things like numpy arrays without specifically checking. There are also functions like https://numpy.org/doc/2.2/reference/generated/numpy.asanyarray.html#numpy.asanyarray which may help with being a little more flexible.

In the long run I suppose it'd be nice to have the low-level functions support multiple values instead of having to iterate over everything every time. I think I tried doing that once but decided it was too hard.

Thoughts?

JaskRendix:
@mraspaud absolutely nothing, I'm going to remove these

djhoese (commented):
Thanks for responding so quickly to the reviews! Do these current changes change the existing return type to any of these methods?

Have you (could you) run mypy on your the files you've changed? I don't think we currently run any type checkers on pyorbital automatically (CI or pre-commit) so it'd be good to know that mypy agrees.

djhoese:
Typically I just go with `mypy <pkgdir>/`, but pyorbital is well behind other pytroll packages on linting/typing standards so we don't have anything existing setup.

JaskRendix:
@djhoese yeah, I create 2 typehints, but I have solved both

```
pyorbital/orbital.py:458: error: Argument 1 to "__call__" of "_UFunc_Nin1_Nout1" has incompatible type "Any | None"; expected "complex | str | bytes | generic[Any]" [arg-type] pyorbital/orbital.py:
464: error: Value of type variable "SupportsRichComparisonT" of "max" cannot be "signedinteger[_32Bit | _64Bit] | Any | None" [type-var]
```
now back to 5
```
(.venv) giorgio@giorgio-Aspire-V3-572G:~/pyorbital$ mypy pyorbital
pyorbital/orbital.py:42: error: Incompatible types in assignment (expression has type "None", variable has type Module)  [assignment]
pyorbital/orbital.py:49: error: Incompatible types in assignment (expression has type "None", variable has type Module)  [assignment]
pyorbital/tests/test_astronomy.py:38: error: Cannot assign to a type  [misc]
pyorbital/tests/test_astronomy.py:38: error: Incompatible types in assignment (expression has type "None", variable has type "type[DataArray]")  [assignment]
pyorbital/geoloc_example.py:29: error: Skipping analyzing "mpl_toolkits.basemap": module is installed, but missing library stubs or py.typed marker  [import-untyped]
pyorbital/geoloc_example.py:29: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
Found 5 errors in 3 files (checked 20 source files)
```

djhoese:
Ok those errors are from ImportError hackery. I'm OK ignore them for now and in the future we can add an inline type ignore comment.

@mraspaud @pnuu do either of you have time to run Satpy's unit tests against this PR and make sure it passes so we're not surprised later?

djhoese:
Unrelated: @JaskRendix this project is planned to be relicensed from GPL to Apache version 2. Once we organize some of the other dependencies we'll ask all past contributor's if they're OK with this in a GitHub issue. This is just a heads up.

djhoese:
Now that pytroll-schedule is cleaned up, do we have any other issues with this PR? Merge?

mraspaud:
I haven't tested it yet, I'll try to get to it tonight

mraspaud (commented):
Thanks a lot for this PR, I appreciate the huge refactoring effort!

I have some comments and questions inline.

mraspaud:
I had an AI agent (claude) look at this and it flagged a few issues. Would you mind having a look at these claims?

Blocking, with evidence

1. get_last_an_time no longer returns nodes on eccentric orbits. The new _find_last_node_time finds a node, then if it's the wrong kind subtracts half an orbital period — which is only valid for a circular orbit — and never re-refines. On AIAA satellite 5 (e=0.186), across 12 request times:

<table>
<tr>
 <td>
 <td>worst distance from the equator at the returned time
<tr>
 <td>main
 <td>0.78 km
<tr>
 <td>PR 206
 <td>1894 km, wrong in 6 of 12 cases
</table>

It returns a plausible-looking timestamp that isn't a node. This is a regression of working behaviour, and eccentric orbits are exactly where a node time is interesting.

2. find_aos/find_aol step over short windows. _find_single_crossing scans in fixed 5-minute steps looking for a sign change. Any visibility window shorter than the step can fall between two samples. Sweeping the search start across one step for a 3.1-minute window above a 40° mask: missed 12 times out of 30 (40%), purely depending on where the grid lands. It returns None, so the caller sees "no pass" rather than an error.

3. Two precision regressions.
- node_func evaluates at int(minutes * 60) seconds, so the node time is quantized to whole seconds — five different request times give one identical answer, where main gives five distinct sub-millisecond ones. It also lands 1.38 km off, outside the original's own ±1 km criterion.
- _get_root now passes rtol=tol to brentq. rtol is relative; with the default tol=0.01 that's 1%. Via get_next_passes(tol=0.001) the rise-time elevation error goes from -0.000000° to -0.000221°. xtol=tol alone would have been the sensible reading of that parameter.

Smaller things

- ecef_to_topocentric and compute_azimuth_elevation are added as public module names to decompose one internal function. That enlarges the API surface permanently for a refactor.
- find_aol's docstring says "Acquisition of Signal (AOL)" — it's Loss of Signal.
- The PR bundles reformatting (black/isort), a demo fix, a refactor, and two new features across 1000 lines. That's most of why it's been open eleven months. Splitting it — the np.clip fix and the __main__ fix are uncontroversial and could land today — would get value in quickly and leave the node/AOS work to be reviewed on its own.

Manny7717 (approved):
Reviewed locally on head 58cbc63 (checked out PR head, ran suites, base-comparison + numeric sweeps). The refactor is behavior-preserving for existing methods and the new functionality is genuinely new (find_aos/find_aol were `pass` stubs, get_last_dn_time didn't exist).

**Verified:**
- **Zero regressions:** the base's original test suite (140 tests, incl. get_observer_look / get_last_an_time / get_next_passes / get_orbit_number expectations) passes 140/140 on HEAD code; head's full suite 190/190 passes.
- **Numeric equivalence:** 20,000-random-geometry sweep of old vs new `get_observer_look` → max azimuth diff 0.0 deg, elevation diff < 1e-9 (the arctan→arctan2 refactor is machine-precision identical on real geometries; the arctan2 form is also the correct fix for the top_s=0 axis case, and the new `clip(-1, 1)` handles el≈-90 where the old code produced NaN).
- **New tests are real:** applied the PR's rewritten test_orbital.py + new test_utils.py to base cd410a9 → 14 failures (crossing-detection cases, find_aos) + collection error (compute_azimuth_elevation missing) — the tests genuinely exercise the new code, and the find_aos/find_aol implementations (sign-change scan + brentq, direction-checked) behave correctly across the parametrized geometry matrix incl. no-pass, extreme-horizon (89°), and below-horizon-altitude cases.
- **get_last_an_time rewrite:** bracket search + brentq root to 0.0001 min (6 ms) is strictly more precise than the old 1 km bisection; ascending/descending disambiguation via v_z at the root is correct (half-period correction direction verified). Base tests confirm identical results.
- **get_next_passes:** same algorithm + a genuine fix — `risetime`/`risemins` are now reset after each completed pass (old code left them set, so a pass in progress at window start could mis-pair); `round()` on `length*60` also makes float lengths safe.
- **No in-place/shared-state issues** in the new helpers; scipy is an existing dependency (brentq), XMNPDA/_get_root pre-exist.

Non-blocking notes:
1. **Fixed 5-min scan step in `_find_single_crossing`** (max 1440 min): a horizon window shorter than 5 min (fast LEO pass with a high horizon threshold, or a rise+set fully inside one step) would be missed and return None. Adaptive/finer step near expected crossings would harden this; the current tests don't cover a sub-5-min window.
2. `get_next_passes`: `or fallmins is None` is dead (fallmins is freshly assigned non-None in that branch) — cosmetic.
3. Commit history is noisy (11+ commits, many "Update tests"); consider squashing before merge.
4. New public methods (find_aos/find_aol/get_last_dn_time) — worth a changelog/release-note entry per repo convention.

Solid PR — the refactor reads cleaner and the behavior tests are meaningful.

JaskRendix:
@mraspaud I don’t know, we’re talking about 3 files, 2 edited and 1 new. And god forbid I’m used to running isort + black before pushing anything. I’m going to close the PR. Feel free to pick whatever you want from it and reuse whatever helps, so things can move faster on your side.

PR #209, opened 2025-10-09
Extended platform support for Kineis Constellation

seapyth0n:
https://www.n2yo.com/database/?name=KINEIS#results

Added platforms to support the Kineis constellation.

25 new IDs were added.

### DIFF
--- a/pyorbital/config.py
+++ b/pyorbital/config.py
@@ -1,0 +1,4 @@
+"""Set up the config object."""
+from donfig import Config
+
+config = Config("pyorbital")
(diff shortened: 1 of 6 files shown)


seapyth0n:
Good catch on the spacing. I updated that in the latest commit, but the codecov is still at 90.35%?

djhoese:
@seapyth0n I'm having trouble finding an email for you, but if you see this could you please review and hopefully approve https://github.com/pytroll/pyorbital/issues/208 based on your contributions here?

PR #224, opened 2026-03-02
Refactor SQLiteTLE and modernize codebase: remove `.format`, drop `(object)`, adopt `pathlib`, and split SQLite tests

JaskRendix:
PR updates the SQLiteTLE class and surrounding code to use modern Python features. It removes `.format()` string formatting, replaces `os.path` with `pathlib` where appropriate, and drops the old‑style `(object)` inheritance. The SQLite test suite is moved out of the general TLE tests into a new file focused on SQLite behavior. The refactor keeps behavior unchanged and all existing tests pass.

- [x] Tests added  
- [x] Fully documented

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -13,15 +13,16 @@
 """Configurations for sphinx based documentation."""
 
 import datetime as dt
-import os
 import sys
+from pathlib import Path
 
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
-# documentation root, use os.path.abspath to make it absolute, like shown here.
+# documentation root, use an absolute path as shown here.
 
-sys.path.insert(0, os.path.abspath("../../"))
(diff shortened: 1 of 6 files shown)


djhoese (approved):
I'm OK with this being merged, but I think my general opinion of the str -> pathlib changes is that they are unnecessary and make some of the code more confusing than clearer. I say that as someone who likes pathlib for new code. There is just a bit too much `str -> Path -> str` conversion here and I'm worried it will introduce bugs (that aren't tested) or will make future functionality more difficult. For example, `Path` doesn't play well with custom protocols if I remember right so if in the future TLEs come from S3 and there is an S3 URI like `s3://tle-bucket/path/to/tle.txt` then converting to `Path` would not work correctly.

```python
In [2]: str(Path("s3://a/b/c/tle.txt"))
Out[2]: 's3:/a/b/c/tle.txt'
```

Note the single slash. So again, I'm fine with this being merged if other maintainers are as well, but we should be careful about wide sweeping "convert to this other API" style changes.

PR #229, opened 2026-04-02
Rework geoloc to support yaw steering and numba for optimisation

mraspaud:
This PR reworks the geoloc module, by providing generic instrument definition builders.

Moreover, we implement yaw-steering and add (optional) numba optimisation.

For generating a full swath of lon/lat, the speed-up goes from 5x to 25x depending on the geometry.

I used AI for this work. 

 - [x] Tests added 
 - [x] Fully documented

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -153,6 +153,105 @@ If we take a TLE from one week earlier we get a slightly different result:
 
 
 
+Instrument geolocation conventions
+----------------------------------
+
+The geolocation code builds a local orbital frame from the satellite position
+and velocity, and rotates the instrument line of sight within it. Two choices in
+that construction affect the ground solution, and both are selectable, because
+changing them moves the geolocation of *existing* products.
+
+**Nadir convention** -- which direction counts as "down":
+
+``"legacy"`` (current default)
+   The direction released versions of Pyorbital have always used: the
+   normalised position of the ellipsoid subpoint of the antipode. The frame is
+   left exactly as released, which means the nadir is *not* re-orthogonalised
+   against the velocity.
+
+``"geocentric"``
+   Straight at the centre of the Earth. Validation of FY-3 MERSI against
+   reference geolocation selects this one: the legacy convention leaves a
+   latitude-dependent error going as ``sin(2 * latitude)``, about 0.3 km at 45
+   degrees for an 850 km orbit and vanishing at the equator.
+
+``"geodetic"``
+   Along the ellipsoid normal. This departs furthest from the other two, by
+   about 2.8 km at 45 degrees, and is *not* what the FY-3 validation supports.
+
+**Rotation order** -- roll and pitch do not commute, and the cross-track field
+of view carries the whole scan angle while pitch is a small attitude bias, so
+the two orders diverge with scan angle: nothing at nadir, growing towards the
+swath edge in proportion to the pitch bias.
+
+``"legacy"`` (current default)
+   Roll first, then pitch, as released. The along-track component of the
+   pointing vector is compressed by ``cos(scan_angle)``.
+
+``"pitch_first"``
+   Pitch first, then roll, so the along-track component is independent of the
+   scan angle. For AVHRR with a fitted pitch bias of 0.16 degrees the two
+   differ by about 1.8 km at the swath edge; for FY-3F, whose bias is five
+   times smaller, by about 0.37 km -- too little for that data to say which is
+   right.
+
+Selecting a convention
+^^^^^^^^^^^^^^^^^^^^^^
+
+Both can be passed directly to :func:`~pyorbital.geoloc.geolocate` and
+:func:`~pyorbital.geoloc.compute_pixels`::
+
+    >>> from pyorbital.geoloc import geolocate
+    >>> lons, lats, alts = geolocate(orbital, scan_geometry, times,   # doctest: +SKIP
+    ...                              nadir_convention="geocentric",
+    ...                              rotation_order="pitch_first")
+
+Libraries built on Pyorbital -- pygac and the readers using it, for instance --
+call the geolocation without forwarding these arguments, so they can also be set
+process-wide through the configuration, which reaches those callers with no
+change to them::
+
+    >>> import pyorbital                                              # doctest: +SKIP
+    >>> with pyorbital.config.set(nadir_convention="geocentric",      # doctest: +SKIP
+    ...                           rotation_order="pitch_first"):
+    ...     reprocess()
+
+or for a whole processing run, through the environment::
+
+    PYORBITAL_NADIR_CONVENTION=geocentric PYORBITAL_ROTATION_ORDER=pitch_first ...
+
+An explicit argument takes precedence over the configuration.
+
+Reproducing existing products
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Leaving both unset keeps the released behaviour, so results do not move
+underneath existing processing chains. Because that behaviour is measurably
+less accurate, relying on the default raises a :exc:`DeprecationWarning`;
+requesting ``"legacy"`` explicitly, or setting it in the configuration, is
+treated as a deliberate choice and is silent.
+
+Pinning both to ``"legacy"`` reproduces released Pyorbital to within floating
+point (4e-11 degrees, well below a micrometre on the ground), which is what an
+archive reprocessing needs::
+
+    >>> with pyorbital.config.set(nadir_convention="legacy",          # doctest: +SKIP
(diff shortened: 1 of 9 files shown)


pyorbital/geoloc.py
```diff
@@ -1,23 +1,42 @@
...
 from pyorbital.orbital import Orbital
 
-A = 6378.137  # WGS84 Equatorial radius (km)
+A = 6378.137  # WGS84 and GRS80 Equatorial radius (km)
 B = 6356.75231414  # km, GRS80
-# B = 6356.752314245 # km, WGS84
+
+# Module-level cached Transformer — avoids re-creating the PROJ context on
+# every get_lonlatalt() call.
+_GEOCENT_TO_LATLONG = None
```
djhoese, line 29: Would it be simpler to do an `@cache` decorator? Essentially the same thing. Or is there really a measurable performance difference in just recreating it?

pyorbital/geoloc.py
```diff
@@ -1,23 +1,42 @@
...
+
+
+def _get_transformer():
+    global _GEOCENT_TO_LATLONG
+    if _GEOCENT_TO_LATLONG is None:
+        _GEOCENT_TO_LATLONG = Transformer.from_crs(
+            dict(proj="geocent"), dict(proj="latlong"))
```
djhoese, line 36: Are there EPSG or pyproj constant versions of these? It seems odd to create these CRSes from the PROJ.4-style dictionary/string rather than something more explicit. Especially if `a` and `b` radii can be specified or use an existing definition that match the `A` and `B` global in this module.

pyorbital/geoloc.py
```diff
@@ -159,52 +335,434 @@ def qrotate(vector, axis, angle):
...
+
 def get_lonlatalt(pos, utc_time):
     """Calculate sublon, sublat and altitude of satellite, considering the earth an ellipsoid."""
-    trf = Transformer.from_crs(dict(proj="geocent"), dict(proj="latlong"))
-    lon, lat, alt = trf.transform(*(pos * 1000))
-    lon = lon - np.rad2deg(astronomy.gmst(utc_time))
+    if _HAS_NUMBA:
+        lon, lat, alt = _numba_ecef_to_lonlatalt(pos[0], pos[1], pos[2])
+    else:
+        lon, lat, alt = _get_transformer().transform(*(pos * 1000))
```
djhoese, line 552: So all of this is to say that the numba version is faster than pyproj? Is that because of the parallel computing part of it or something else in the way the algorithm is implemented?

mraspaud: There is no magic in the implemented algorithm, so just the parallelism + jit made it much faster than pyproj

djhoese: Well pyproj and PROJ are both compiled so I'd assume it is purely the parallelism. I'm mostly worried about numba as a dependency when depending on Cython and openmp (using `prange`) may be "easier" in terms of getting this optimization all the time or at least in more instances/environments since openmp is also often required for pykdtree optimizations. There would be no conditionals of which function is used, it would just be a C version. Or...dask could be an option too since it is still parallel but with an overhead. Obviously you've already implemented this solution so it is hard to not use it and implement something else.

pyorbital/geoloc.py
mraspaud: Fixed in [260b554](https://github.com/pytroll/pyorbital/pull/229/commits/260b55476bc2fa6aadea3ee3df37601921374a38)

pyorbital/geoloc.py
mraspaud: Fixed in [260b554](https://github.com/pytroll/pyorbital/pull/229/commits/260b55476bc2fa6aadea3ee3df37601921374a38) along with refactored in other places too (using constants)

pyorbital/geoloc.py
mraspaud: Well, it's in place now, and anyway the numba requirement is optional, so we can revert do python if needed. If something breaks with cython or omp, I can fix it, but I didn't see any issues in my tests.

pyorbital/geoloc.py
```diff
@@ -26,17 +27,24 @@
...
-    return _GEOCENT_TO_LATLONG
+    """Get a geocentric to lon/lat transformer.
+
+    The ellipsoid is taken from the module-level *A* and *B* constants, which the
+    numba kernels read too, so every path in this module shares one declaration.
+    The result is cached to avoid re-creating the PROJ context on every call.
+    """
+    return Transformer.from_crs(
+        dict(proj="geocent", a=A * 1000, b=B * 1000),
+        dict(proj="latlong", a=A * 1000, b=B * 1000))
```
djhoese, line 38: Ok, yeah this is more exact. I was more thinking of something like:

https://pyproj4.github.io/pyproj/stable/api/crs/crs.html#geocentriccrs

But was hoping for something even better like `pyproj.GEOCENTRIC` or some constant CRS. I'm not sure that exists though.

doc/source/index.rst
```diff
@@ -153,6 +153,105 @@ If we take a TLE from one week earlier we get a slightly different result:
...
+that construction affect the ground solution, and both are selectable, because
+changing them moves the geolocation of *existing* products.
+
+**Nadir convention** -- which direction counts as "down":
+
+``"legacy"`` (current default)
+   The direction released versions of Pyorbital have always used: the
+   normalised position of the ellipsoid subpoint of the antipode. The frame is
+   left exactly as released, which means the nadir is *not* re-orthogonalised
+   against the velocity.
```
djhoese, line 170: Is there any name for this method outside of "legacy" or "old school pyorbital"? Something we can use that others outside of pyorbital might recognize? Are there numbers on how far this method is from what's considered correct like the other methods document? Or are the other methods using this as a reference?

pyproject.toml
```diff
@@ -5,13 +5,14 @@ description = "Scheduling satellite passes in Python"
...
 ]
-dependencies = ["numpy>=1.19.0",
-                "scipy",
-                "requests",
-                "defusedxml",
-                "pyproj>=3.7.1",
-                "donfig",
-                ]
+dependencies = [
+    "numpy>=1.19.0",
```
djhoese: Might as well put this in here, right?

```suggestion
    "numpy>=2.0.0",
```

pyorbital/geoloc_instrument_definitions.py
djhoese, line 1: Just a thought after skimming and refusing to read all these lines of code that I'm not previously familiar with...should there be a `geoloc` subpackage with `geoloc.py` and `geoloc_instrument_definitions.py` in it?

pyorbital/astronomy.py
```diff
@@ -159,6 +159,17 @@ def sun_zenith_angle(utc_time, lon, lat):
     return np.rad2deg(np.arccos(cos_zen(utc_time, lon, lat)))
 
 
+def sun_azimuth_angle(utc_time, lon, lat):
```
djhoese, line 162: Did this exist somewhere else that I'm not seeing? Type annotations useful or is it so generic that it wouldn't help much?

djhoese (approved):
This looks like an all around huge improvement. I had a couple comments inline, but relatively minor things. I do fear that there are a lot of new tests that could maybe be combined (AI likes adding lots of tests), but I'm hoping you did a first pass at trying to combine them. I did not sift through most of the test modules or the two geoloc modules in detail this time as I really don't understand what they're doing.

doc/source/index.rst
mraspaud: "bad_version"? "wrong_nadir"? ;) seriously, I can’t really think of something else…

pyorbital/geoloc_instrument_definitions.py
mraspaud: Good idea, but how about we do that in another PR?

pyorbital/astronomy.py
mraspaud: No, there wasn’t. Usually we would just call get_alt_az and pick the second return item.

pyorbital/geoloc.py
mraspaud: There are no pyproj constants for these afaics. We could use GeocentricCRS(), but I’d rather keep them as explicit constants in the module, if that’s ok with you.

djhoese: Yep, all good.

pyproject.toml
mraspaud: addressed in [caf82e9](https://github.com/pytroll/pyorbital/pull/229/commits/caf82e914b090a3d1ae5018b40be82c4138af200) (along with another pass on the tests)

PR #156, opened 2024-06-26
Fix dtype preservation in astronomy functions

djhoese:
Redo of #155 without type annotation

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [x] Tests passed 
 - [x] Passes ``flake8 pyorbital`` 
 - [x] Fully documented

PR #238, opened 2026-07-20
Fix sun_zenith_angle on integer lon/lat (crash + silent truncation)

binggao1230:
`sun_zenith_angle` reads `.dtype` off the raw `lon`/`lat` argument, so an integer scalar raises `AttributeError` and an integer array is silently truncated to whole degrees (~1° error). The sibling functions (`get_alt_az`, `cos_zen`, `observer_position`) avoid this because `np.deg2rad` promotes int→float before `.dtype` is read.

```python
>>> import datetime as dt, numpy as np
>>> from pyorbital import astronomy
>>> t = dt.datetime(2020, 6, 21, 12)
>>> astronomy.sun_zenith_angle(t, 10, 50)
AttributeError: 'int' object has no attribute 'dtype'
>>> astronomy.sun_zenith_angle(t, np.array([10, 20, 30]), np.array([50, 55, 60]))
array([27, 34, 41])          # truncated; float input gives [27.586, 34.733, 41.957]
```

The recast is redundant — `cos_zen` already restores the input precision — so removing it fixes both cases while preserving float32/float64 dtypes. This is a residual of the dtype-preservation idiom from #156: that site delegates the angle math to `cos_zen` instead of calling `np.deg2rad` itself, so it read the un-promoted argument. Adds the missing integer-input test coverage; `test_astronomy.py` stays green (21 → 27 passed).

### DIFF
--- a/pyorbital/astronomy.py
+++ b/pyorbital/astronomy.py
@@ -155,10 +155,8 @@ def sun_zenith_angle(utc_time, lon, lat):
     lon,lat in degrees.
     The sun zenith angle returned is in degrees.
     """
-    sza = np.rad2deg(np.arccos(cos_zen(utc_time, lon, lat)))
-    if not isinstance(lon, float):
-        sza = sza.astype(lon.dtype)
-    return sza
+    # cos_zen already restores the input dtype, so no separate recast is needed here.
+    return np.rad2deg(np.arccos(cos_zen(utc_time, lon, lat)))
 
 
 def sun_earth_distance_correction(utc_time):
--- a/pyorbital/tests/test_astronomy.py
+++ b/pyorbital/tests/test_astronomy.py
@@ -93,6 +93,36 @@ class TestAstronomy:
             np.testing.assert_allclose(sun_theta, exp_theta, atol=abs_tolerance)
             assert isinstance(sun_theta, type(lon))
 
+    @pytest.mark.parametrize(
+        ("lon", "lat"),
+        [
+            (10, 50),
+            (0, 0),
+            (16, 59),
+            (-75, -30),
+        ]
+    )
+    def test_sun_zenith_angle_integer_scalar(self, lon, lat):
+        """Test that an integer scalar lon/lat neither crashes nor truncates to whole degrees."""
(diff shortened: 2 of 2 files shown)


djhoese:
Thanks for this. It is hard to argue with the test results. I wonder if this logic was added when dask and numpy didn't play as nicely with dtype casting and are little more sane now. @mraspaud or @pnuu or @adybbroe want to sanity check this with another review?

djhoese (approved):

pyorbital/astronomy.py
```diff
@@ -155,10 +155,8 @@ def sun_zenith_angle(utc_time, lon, lat):
     lon,lat in degrees.
     The sun zenith angle returned is in degrees.
     """
-    sza = np.rad2deg(np.arccos(cos_zen(utc_time, lon, lat)))
-    if not isinstance(lon, float):
-        sza = sza.astype(lon.dtype)
-    return sza
+    # cos_zen already restores the input dtype, so no separate recast is needed here.
```
Manny7717, line 158: Confirmed the claim in this comment: numpy ufunc dtype rules make the explicit recast redundant — float32 stays float32, ints promote to float64. Removing the `astype(lon.dtype)` also fixes the silent int-truncation case. Good catch on the root cause.

Manny7717 (approved):
## Review — Approved

Verified locally: the new `test_sun_zenith_angle_integer_scalar` and `test_sun_zenith_angle_integer_array` tests pass, and integer inputs now return float angles matching the float64 reference to 1e-8.

I also checked the dtype-preservation concern that the removed recast was guarding: `cos_zen` runs through numpy ufuncs (`deg2rad`, `sin`, `cos`, `arccos`), which preserve float32 input dtype for float inputs and promote integer inputs to float64 — so float32 arrays still come back float32, int scalars no longer crash with `AttributeError: 'int' object has no attribute 'dtype'`, and int arrays are no longer silently truncated to whole degrees. The simplification is safe.

Note: 8 `test_sunangles` cases (xarray numpy/dask variants) fail in my environment, but they fail **identically on main** — pre-existing env/version noise unrelated to this PR.

mraspaud (approved):
LGTM

PR #242, opened 2026-08-28
fix: apply a bounded timeout to TLE downloads

### LINKED ISSUE #218: `Orbital("Landsat-8")` call hangs indefinitely on first invocation
I am encountering an issue where `Orbital("Landsat-8")` sometimes hangs indefinitely. What could be the underlying cause of this intermittent hanging during the initial call?
## Steps to reproduce:
1. In a script, I call:
```python
Orbital("Landsat-8")
```
2. On the first run, this call often hangs forever without raising any error or warning.
3. I interrupt the process with `Ctrl+C`.
4. Then on subsequent runs, it usually works, though sometimes it still requires multiple attempts.
----
OS: Fedora
Pyorbital Version: 1.9.2

Manny7717:
## What

Fixes #218: `Orbital("Landsat-8")` (or any satellite) can hang **forever** on the first invocation when no local TLE file is configured.

## Problem

The default internet TLE path calls `urllib.request.urlopen(url)` **without a timeout** (`pyorbital/tlefile.py`):

- `fetch()` — line 139
- `_get_internet_uris_and_open_method()` — returns bare `urlopen` as the open function for `Tle._read_tle()`

A stalled Celestrak connection (throttling, DNS hiccup, dropped SYN) blocks indefinitely with no error or warning — exactly the intermittent hang in the report. Ctrl+C "fixes" it because the OS/socket layer recovers, so subsequent runs work. The newer `Downloader.fetch_plain_tle` path already bounds requests with `requests.get(uri, timeout=15)`; the legacy `urlopen` path was never given the same bound.

## Fix

Add a small `_urlopen_with_timeout(url)` helper that calls `urlopen(url, timeout=15)` — the same 15-second bound `fetch_plain_tle` uses — and route both `fetch()` and `_get_internet_uris_and_open_method()` through it. A stalled connection now raises `URLError` after 15 s instead of hanging the caller forever. No API or behavior change for the local-file paths.

## Tests

- `test_internet_tle_fetch_applies_timeout` — asserts the internet open function passes `timeout=15` to `urlopen`. **Fails on the pre-fix code** (passes `timeout=None`).
- `test_fetch_fails_loudly_on_stalled_connection` — asserts a stalled connection raises `URLError` instead of blocking.
- Full suite: 147 passed (44 in `test_tlefile.py`).

## Limitations

The 15 s bound matches the existing `fetch_plain_tle` convention; it is not configurable. DNS resolution itself is bounded by the OS resolver, not by this timeout.

### DIFF
--- a/pyorbital/tests/test_tlefile.py
+++ b/pyorbital/tests/test_tlefile.py
@@ -889,7 +889,7 @@ def test_tle_instance_printing():
     assert str(tle) == expected
 
 
-def test_internet_tle_fetch_applies_timeout(monkeypatch):
+def test_internet_tle_fetch_applies_timeout():
     """Regression test for #218: internet TLE downloads must use a bounded timeout.
 
     A stalled Celestrak connection used to hang ``Orbital(...)`` forever on the
@@ -898,28 +898,35 @@ def test_internet_tle_fetch_applies_timeout(monkeypatch):
     """
     from pyorbital import tlefile
 
-    calls = []
+    with mock.patch.object(tlefile, "urlopen") as fake_urlopen:
+        fake_urlopen.return_value = io.StringIO("")
 
-    def _fake_urlopen(url, timeout=None):
-        calls.append((url, timeout))
-        return io.StringIO("")
+        _, open_func = tlefile._get_internet_uris_and_open_method()
+        open_func("https://celestrak.example/group")
 
-    monkeypatch.setattr(tlefile, "urlopen", _fake_urlopen)
-
-    _, open_func = tlefile._get_internet_uris_and_open_method()
-    open_func("https://celestrak.example/group")
-    assert calls == [("https://celestrak.example/group", 15)]
+    fake_urlopen.assert_called_once_with(
+        "https://celestrak.example/group", timeout=15)
 
 
-def test_fetch_fails_loudly_on_stalled_connection(monkeypatch, tmp_path):
+def test_tle_fetch_timeout_configurable_via_env():
+    """PYORBITAL_TLE_FETCH_TIMEOUT overrides the default TLE fetch timeout."""
+    from pyorbital import tlefile
+
+    with mock.patch.dict(os.environ, {"PYORBITAL_TLE_FETCH_TIMEOUT": "42"}), \
+            mock.patch.object(tlefile, "urlopen") as fake_urlopen:
+        fake_urlopen.return_value = io.StringIO("")
+        tlefile._urlopen_with_timeout("https://celestrak.example/group")
+
+    fake_urlopen.assert_called_once_with(
+        "https://celestrak.example/group", timeout=42.0)
+
+
+def test_fetch_fails_loudly_on_stalled_connection(tmp_path):
     """A stalled TLE download raises instead of hanging the caller forever."""
     from pyorbital import tlefile
 
-    def _stalled(url, timeout=None):
-        raise urllib.error.URLError("timed out")
-
-    monkeypatch.setattr(tlefile, "urlopen", _stalled)
-
-    destination = tmp_path / "tles.txt"
-    with pytest.raises(urllib.error.URLError):
-        tlefile.fetch(str(destination))
+    with mock.patch.object(
+            tlefile, "urlopen", side_effect=urllib.error.URLError("timed out")):
+        destination = tmp_path / "tles.txt"
+        with pytest.raises(urllib.error.URLError):
+            tlefile.fetch(str(destination))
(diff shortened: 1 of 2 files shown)


pyorbital/tlefile.py
```diff
@@ -30,6 +30,16 @@
 TLE_URLS = [f"https://celestrak.org/NORAD/elements/gp.php?GROUP={group}&FORMAT=tle"
             for group in TLE_GROUPS]
 
+# Timeout for TLE downloads from the internet (seconds). Kept in sync with the
+# timeout used by Downloader.fetch_plain_tle so a stalled connection fails
+# loudly instead of hanging the caller forever.
```
mraspaud: Unnecessary comment, please remove.

mraspaud (changes requested):
Thanks for fixing this. One comment inline.

Manny7717:
Thanks for the review! Removed the redundant comment above `_TLE_FETCH_TIMEOUT` (commit 8439462). tlefile tests still 44/44 passing.

pyorbital/tests/test_tlefile.py
```diff
@@ -885,3 +887,39 @@ def test_tle_instance_printing():
     expected = "{'arg_perigee': 130.536,\n 'bstar': -1.1606e-05,\n 'classification': 'U',\n 'eccentricity': 0.0006703,\n 'element_number': 292,\n 'ephemeris_type': 0,\n 'epoch': np.datetime64('2008-09-20T12:25:40.104192'),\n 'epoch_day': 264.51782528,\n 'epoch_year': '08',\n 'id_launch_number': '067',\n 'id_launch_piece': 'A  ',\n 'id_launch_year': '98',\n 'inclination': 51.6416,\n 'mean_anomaly': 325.0288,\n 'mean_motion': 15.72125391,\n 'mean_motion_derivative': -2.182e-05,\n 'mean_motion_sec_derivative': 0.0,\n 'orbit': 56353,\n 'right_ascension': 247.4627,\n 'satnumber': '25544'}"  # noqa
 
     assert str(tle) == expected
+
+
+def test_internet_tle_fetch_applies_timeout(monkeypatch):
```
djhoese: We generally use unittest's mock. Would you mind rewriting these to use that instead of monkeypatch? My understanding is that mixing the two can have some bad consequences sometimes. Plus AI seems to lean towards monkeypatch if it sees it anywhere in the codebase.

@mraspaud feel free to override.

pyorbital/tlefile.py
```diff
@@ -30,6 +30,13 @@
 TLE_URLS = [f"https://celestrak.org/NORAD/elements/gp.php?GROUP={group}&FORMAT=tle"
             for group in TLE_GROUPS]
 
+_TLE_FETCH_TIMEOUT = 15
```
djhoese: Should this be configurable with an environment variable?

Manny7717:
Thanks for the review! Both points addressed in f026868:

1. **Tests rewritten with `unittest.mock`** — `test_internet_tle_fetch_applies_timeout` and `test_fetch_fails_loudly_on_stalled_connection` now use `mock.patch.object(tlefile, "urlopen")` instead of `monkeypatch`, matching the file's existing `from unittest import mock` import style.

2. **Timeout is now configurable via `PYORBITAL_TLE_FETCH_TIMEOUT`** — `_get_tle_fetch_timeout()` reads the env var (float), falls back to the 15s default when unset or invalid (with a warning log). `_urlopen_with_timeout` uses it, so both `fetch()` and `_get_internet_uris_and_open_method()` honor it. Added `test_tle_fetch_timeout_configurable_via_env` covering the override.

All 45 tlefile tests pass locally (44 pre-existing + the new env-var test).

pyorbital/tlefile.py
```diff
@@ -30,6 +30,26 @@
...
+
+
+def _get_tle_fetch_timeout():
+    """Get the TLE fetch timeout, configurable via ``PYORBITAL_TLE_FETCH_TIMEOUT``."""
+    raw = os.environ.get("PYORBITAL_TLE_FETCH_TIMEOUT")
```
djhoese, line 36: What do you think about removing `_TLE_FETCH_TIMEOUT` and just making a default variable locally in this function that is set to the same thing? I don't see a need for a global variable.

Manny7717: Done — `_TLE_FETCH_TIMEOUT` is now a local `default_timeout = 15` inside `_get_tle_fetch_timeout()` (commit 1de1fb7). The env override + invalid-value fallback behavior is unchanged; 45/45 tlefile tests pass. Thanks for the review!

PR #243, opened 2026-08-31
Propagate deep-space satellites (SDP4)

mraspaud:
# What this does

pyorbital refused any satellite whose orbit takes 225 minutes or more, and also refused near-earth satellites with a low enough perigee to need the simplified drag equations. Between them that was 26 of the 29 satellites in the AIAA verification set. This adds the deep-space model from AIAA 2006-6753 in a new module, pyorbital/_sgdp4_deep.py, hooked into the existing propagator: the geometry of the Sun and Moon at the epoch, the slow drift they impose, the periodic wobble they impose, and — for orbits that circle once or twice while the Earth turns once — the beat against the Earth's own equatorial bulges, integrated forward in twelve-hour steps.

All 29 verification satellites now propagate, with no skips.

# Two bugs found on the way

The mean motion constant was rounded. XKE was hardcoded as 0.743669161e-1, the nine-digit decimal printed in Spacetrack Report #3. Its relative error of 4.5e-10 lands in the semi-major axis, shifts the computed perigee by 1.9e-6 km, and because the atmospheric drag fit takes (120 - sfour) — a difference of nearly equal altitudes — to the fourth power, it comes out amplified 370× for low-perigee satellites. It cost half a metre of along-track position over a day and kept three satellites from being propagated at all. XKE and QOMS2T are now computed from the defining WGS-72 parameters. This improved every near-earth satellite by roughly 100×, including ones that were already passing.

The drag correction was applied where it shouldn't be. The mean-anomaly and argument-of-perigee correction was applied unconditionally; Vallado applies it only when the full drag expansion is in use. Deep-space and low-perigee satellites share the simplified equations, so that condition is now named rather than tested in four places.

# Also

- Deep-space satellites propagate to an array of times, ~150× faster than looping.
- Propagating to a set of times now gives bit-for-bit what propagating to each gives. Kepler's equation was iterated until every time in the array had converged, so the ones that converged first kept being nudged and their last bits depended on what else was in the array.
- The resonance integrator restarts from the epoch on every call rather than carrying state, so an answer never depends on what was asked before. Cost is linear in distance from epoch; an array shares one march.
- Removed a wrapper class of 40 properties that only forwarded to the object beside it.
- Verification harness rewritten as one test per satellite, reaching the propagator through the public API only, with a meta-test that makes a silently-skipped satellite impossible.

# Accuracy

All 29 satellites match aiaa_results within main's tolerances (5e-6 km, 5e-9 km/s). Worst position error is 4.475e-06 km on satellite 23333 (WIND, eccentricity 0.973) — 89% of the budget. Worst velocity error is 45% of budget. Beyond the verification set, ~24,000 propagated states across ~2,270 synthetic orbits spanning inclination, eccentricity and mean motion agree with the reference implementation to under 1e-5 km for every orbit whose perigee is above ground.

# Known follow-ups

1. Satellite 23333 uses 89% of the position tolerance. It's newly under test (previously skipped as deep-space), and the reference implementation matches the file to 4.6e-9 km where we're at 4.5e-6. A real difference specific to extreme eccentricity, not yet explained — the Kepler stopping criterion accounts for only a small part of it. Worth a look before anyone tightens tolerances further.
2. Mutation testing showed the verification set doesn't pin three branches (an eccentricity fit boundary, a resonance band edge, and the geosynchronous eccentricity terms). test_deep_space_branches.py adds three orbits that do.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [x] Fully documented

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ pip-log.txt
 .coverage
 .tox
 nosetests.xml
+.benchmarks
 
 # Translations
 *.mo
--- a/pyorbital/_sgdp4_deep.py
+++ b/pyorbital/_sgdp4_deep.py
@@ -47,12 +47,64 @@ class DeepSpace:
     """
 
     def __init__(self, days_since_1949, eccentricity, arg_perigee, inclination,
-                 right_ascension, mean_motion):
+                 right_ascension, mean_motion, mean_anomaly, sidereal_time_at_epoch,
+                 mean_anomaly_rate, arg_perigee_rate, right_ascension_rate,
+                 semi_major_axis):
         """Work out the geometry and the coefficients that follow from it."""
         self._geometry = compute_lunisolar_geometry(
             days_since_1949, eccentricity, arg_perigee, inclination, right_ascension,
             mean_motion)
         self._secular_rates = compute_secular_rates(self._geometry)
+        self._mean_motion_at_epoch = mean_motion
+        self._arg_perigee_at_epoch = arg_perigee
+        self._arg_perigee_rate = arg_perigee_rate
+        self.resonance = classify_resonance(mean_motion, eccentricity)
+        if self.resonance != NOT_RESONANT:
+            self._prepare_resonance(eccentricity, arg_perigee, right_ascension,
+                                    mean_anomaly, sidereal_time_at_epoch, mean_motion,
+                                    mean_anomaly_rate, arg_perigee_rate,
+                                    right_ascension_rate, semi_major_axis)
+
+    def _prepare_resonance(self, eccentricity, arg_perigee, right_ascension, mean_anomaly,
+                           sidereal_time_at_epoch, mean_motion, mean_anomaly_rate,
+                           arg_perigee_rate, right_ascension_rate, semi_major_axis):
+        """Work out how strongly the orbit beats against the Earth, and from where."""
+        rates = self._secular_rates
+        aonv = 1.0 / semi_major_axis
+        theta = sidereal_time_at_epoch
+
+        if self.resonance != RESONANT_TWICE_A_DAY:
+            raise NotImplementedError(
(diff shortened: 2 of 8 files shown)


Manny7717 (approved):
Verified locally on head b582f15 (AIAA 2006-6753 acceptance suite + full test run):

- **AIAA verification:** 65/65 pass (test_aiaa.py + test_deep_space_branches.py) — every satellite in SGP4-VER.TLE propagated to every reference state within the declared tolerances (5e-6 km position / 5e-9 km/s velocity), including the once-a-day, twice-a-day, and non-resonant deep-space branches, the simplified-drag near-earth branch, and the checksum-error satellites. The acceptance suite structure is sound:  guards against silently dropping cases.
- **No regressions:** full pyorbital/tests run → 12 failed / 201 passed on head vs 12 failed / X passed on base c6a2120 — **byte-identical failure sets** (all pre-existing xarray/dask env noise in TestGetObserverLook*, present on base too). Head additionally passes the 65 new verification tests + new test_orbital.py cases.
- **Code review:** new  cleanly separated from the propagator core; DeepSpace/apply_periodics wiring in _SGDP4 gated on the period threshold; epoch handling and the 2005-12-31 leap-second discontinuity explicitly accounted for in the test harness (one-second offset between reference calendar times and epoch-relative minutes). geoloc.py discriminant clamp (commit 35264db) is a sound, documented behavior fix: negative discriminant → closest-approach point instead of NaN, keeping results finite and continuous for off-Earth attitude sweeps.
- **Design notes (non-blocking):** this is a large rewrite (417 lines in orbital.py) of a numerically sensitive module; the AIAA reference suite is exactly the right safety net, and it passes. The mantissa-level AT_ONCE_TOLERANCE for vectorized-vs-scalar agreement is pragmatic.

Clean, well-tested, and reference-verified. Approving.

PR #244, opened 2026-08-31
Propagate orbits of no eccentricity at all

mraspaud:
A circle is a perfectly good orbit, and the model propagates one as well as it does any other, but the element set was turned away before it got that far: the eccentricity had to be greater than zero, not merely not less than it.

Letting it through, a circular orbit agrees with the reference implementation to a ten-thousandth of a millimetre over two hundred and ten states, spread across inclinations from a thousandth of a degree to retrograde and periods from a low orbit to a geosynchronous one. The reference implementation never objected to these element sets in the first place.

The test propagates such an orbit for one revolution and checks that its distance from the centre of the Earth stays put, wandering only the fifteen kilometres the equatorial bulge accounts for.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented

### DIFF
--- a/pyorbital/orbital.py
+++ b/pyorbital/orbital.py
@@ -1272,7 +1272,9 @@ class _Keplerians:
 
 
 def _check_orbital_elements(orbit_elements):
-    if not (0 < orbit_elements.eccentricity < ECC_LIMIT_HIGH):
+    # An orbit of no eccentricity at all is a circle, which is a perfectly good
+    # orbit and one this model propagates as well as any other.
+    if not (0 <= orbit_elements.eccentricity < ECC_LIMIT_HIGH):
         raise OrbitalError("Eccentricity out of range: %e" % orbit_elements.eccentricity)
     if not ((0.0035 * 2 * np.pi / XMNPDA) < orbit_elements.original_mean_motion < (18 * 2 * np.pi / XMNPDA)):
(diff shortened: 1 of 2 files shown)


Manny7717 (approved):
Verified locally (checkout head d1125ac, venv from repo):

- **Regression proven:** applying the new test to base 3db4cd8 → `test_a_circular_orbit_can_be_propagated` FAILS with `OrbitalError: Eccentricity out of range: 0.000000e+00` at orbital.py:1276 (the `0 < ecc` strict check rejects e=0 before propagation); PASSES on head with the `0 <= ecc` change. The radius-stability assertions (max-min < 20 km, mean ≈ 7205 km) hold on the circular TLE — real propagation, not just constructor acceptance.
- **Zero regressions:** full suite head = 140 passed / 12 failed vs base = 139 passed / 12 failed (same 12 pre-existing xarray/dask sunangles + GetObserverLook env failures, byte-identical sets) — the only difference is the new test itself failing on base, i.e. exactly the intended behavior change.
- Fix is minimal and correct: e=0 is a valid circular orbit; `ECC_LIMIT_HIGH` upper bound unchanged. No other callers of `_check_orbital_elements` affected.

Clean, well-tested maintainer fix. Approving.

djhoese (approved):

PR #245, opened 2026-09-02
Add the METimage instrument definition

mraspaud:
METimage (VII) on Metop-SG-A is a whisk-broom scanner recording 24 lines per sweep of its rotating telescope, so it fits MultiLineWhiskbroomScan directly.  The geometry comes from the EPS-SG VII L1B product format specification (3144 across-track samples, 24 along-track samples per scan) and from the instrument description (1.729 s per revolution at constant scan angle, 0.6 mrad IFOV, 108 degrees of Earth view).

Without this, trollsched fails with an AttributeError when computing a METimage swath boundary, as it looks the instrument up by name in pyorbital.geoloc_instrument_definitions.

 - [x] Tests added

### DIFF
--- a/pyorbital/geoloc_instrument_definitions.py
+++ b/pyorbital/geoloc_instrument_definitions.py
@@ -786,6 +786,84 @@ def viirs_edge_geom(scans_nb):
 
 ################################################################
 #
+#   METimage (VII)
+#
+################################################################
+
+# METimage (also known as VII, the VIS/IR Imager) on Metop-SG-A.
+#
+# Sources:
+#   EPS-SG VII Level 1B Product Format Specification,
+#   EUM/LEO-EPSSG/SPE/14/777138 v5A (20 September 2024), Table 10:
+#   https://user.eumetsat.int/s3/eup-strapi-media/EPS_SG_VII_Level_1_B_Product_Format_Specification_654c0b397a.pdf
+#     num_pixels = 3144 (across-track samples per scan)
+#     num_pixels_alt = 24 (along-track samples per scan)
+#     Pixel 1 is the furthest point from nadir on the left side with respect
+#     to the spacecraft motion.
+#
+#   C. Cao, "Introducing METImage: EUMETSAT's next generation polar imager on
+#   METOP-SG", NOAA STAR seminar, 21 August 2019:
+#   https://www.star.nesdis.noaa.gov/star/documents/seminardocs/2019/20190821_Cao.pdf
+#     Whisk-broom scanner, 24 detector lines per scan, 1.729 s per scan
+#     rotation at constant scan angle, IFOV 0.6 mrad, and a scan mirror
+#     covering an extended Earth view of 108 degrees per revolution.
+#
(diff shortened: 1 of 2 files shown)


Manny7717 (approved):
Verified on head 206d42e vs base origin/main 3840769.

**Purpose confirmed.** `geoloc_instrument_definitions` previously had no METimage entry, so trollsched's swath-boundary lookup by instrument name would AttributeError (as the PR body states). The new `metimage()`/`metimage_edge_geom()` + `METIMAGE_SCAN` (MultiLineWhiskbroomScan) close that gap.

**Geometry is internally consistent and matches the cited sources.** Numbers cross-check: 108° Earth view / 3144 pixels = 0.6 mrad/pixel (matches the documented IFOV); scan half-angle (pixels−1)/2 steps ≈ 53.98°; pixel dwell time = sweep time / pixels where sweep time = 1.729 s × 108/360 (constant scan rate); along-track step = same sampling step (square pixels, 24 lines/scan). All are documented with the EPS-SG VII spec + Cao/Bruens references in the header.

**Regression-proven.** The 6 new tests (shape, ±54° swath edges, edge-to-edge 108°, 0.6 mrad IFOV cross-check, scan timing 108/360 fraction + full-revolution scan spacing, and a real geolocation of the edge geometry with a Metop-A TLE landing ~2800 km swath at zero altitude) all FAIL on base (6 failed — no metimage symbol) and PASS on head (6 passed).

**Zero regressions.** Full test_geoloc.py: head 115 passed / 1 skipped vs base 109 passed / 1 skipped — delta is exactly the 6 new tests, identical skip, no failures on either side. ruff (v0.16.1, repo pre-commit pin) clean on both changed files.

Non-blocking notes: (1) `metimage_edge_geom` hardcodes `np.array([0, _METIMAGE_PIXELS - 1])` instead of reading pixel 1 as "left" via the scan's own convention, but the sign convention is asserted in the tests (cross_track[0] > 0) so drift would be caught; (2) the module-level scan objects pattern matches the existing `viirs`/`modis` style, so no new pattern introduced.

TAlonglong:
can you also add the alias vii? If you get the instrument from the filename you will get vii.

pyorbital/geoloc_instrument_definitions.py
```diff
@@ -784,6 +784,84 @@ def viirs_edge_geom(scans_nb):
...
+
+METIMAGE_SCAN = MultiLineWhiskbroomScan(
+    pixels_per_scan=_METIMAGE_PIXELS,
+    scan_angle=_METIMAGE_SCAN_ANGLE,
+    scan_rate=_METIMAGE_SCAN_RATE,
+    pixel_dwell_time=_METIMAGE_SWEEP_TIME / _METIMAGE_PIXELS,
+    lines_per_scan=_METIMAGE_LINES_PER_SCAN,
+    along_track_step=_METIMAGE_SAMPLING_STEP,
+)
+
```
TAlonglong, line 844: Add vii as an alias.

Maybe vii = metimage wil work

TAlonglong (changes requested):

PR #246, opened 2026-09-02
Add get_last_dn_time, find_aos and find_aol

### LINKED ISSUE #95: Would be nice to have a `get_last_dn_time` to get the last descending node

mraspaud:
Picks up the two features from #206, which the author closed and invited us to
salvage. Both are written fresh; none of that branch's code is carried over.

`get_last_dn_time` gives the last descending node. It shares the existing
ascending node search, which already knew how to walk back to an equator
crossing and bisect onto it — a descending node is the same search with the
satellite heading the other way, so the two now differ only by a sign.

`find_aos` and `find_aol` were stubs that did nothing. `find_aos` asks
`get_next_passes` and takes the rise time of the first pass. `find_aol` scans
for the setting itself, because it has to answer about a pass already under
way and `get_next_passes` discards those.

Two things worth knowing about the result:

- Called mid-pass, the two describe **different passes**. `find_aos` answers
  about the next pass, `find_aol` about the end of the current one. That is
  what you want when you ask "when do I lose this satellite?", but it means
  they don't pair up. Both docstrings say so.
- Elevation is sampled once a minute, so a pass that begins and ends between
  two samples is not seen. #206 sampled every five minutes and missed short
  passes outright; this makes that rarer, not impossible. Fixing it properly
  means a step derived from the pass duration, which is separate work.

Asking for a horizon the satellite never reaches used to give a bare
`IndexError` or a silent `None`. Both now say which satellite, which horizon
and which window came up empty.

 - [x] Closes #95
 - [x] Tests added
 - [x] Tests passed
 - [x] Fully documented

### DIFF
--- a/pyorbital/orbital.py
+++ b/pyorbital/orbital.py
@@ -255,16 +255,15 @@ class Orbital(object):
         """
         passes = self.get_next_passes(utc_time, _HORIZON_SEARCH_HOURS, lon, lat, alt, horizon=horizon)
         if not passes:
-            raise ValueError(f"{self.satellite_name} does not rise above {horizon} degrees "
-                             f"for this observer within {_HORIZON_SEARCH_HOURS} hours of {utc_time}")
+            raise self._no_crossing_found(horizon, utc_time, rising=True)
         rise_time, _, _ = passes[0]
         return rise_time
 
     def find_aol(self, utc_time, lon, lat, alt=0, horizon=0):
         """Find when the satellite next sets below the observer's horizon.
 
-        The search runs over the next 24 hours, and returns None if the
-        satellite does not set in that time. Unlike find_aos, a pass already
+        The search runs over the next 24 hours, and raises a ValueError if no
+        setting is found in that time. Unlike find_aos, a pass already
         under way when *utc_time* falls inside it does count: the answer is
         then the end of that pass, so the two need not describe the same one.
 
@@ -279,6 +278,13 @@ class Orbital(object):
                 minutes = _get_root(elevation_at, crossing, crossing + 1.0,
                                     tol=_CROSSING_TOLERANCE_SECONDS / 60.0)
                 return utc_time + dt.timedelta(minutes=minutes)
+        raise self._no_crossing_found(horizon, utc_time, rising=False)
+
+    def _no_crossing_found(self, horizon, utc_time, rising):
+        """Explain that the horizon crossing asked for is not in the searched window."""
+        crossing = "rise above" if rising else "set below"
+        return ValueError(f"{self.satellite_name} does not {crossing} {horizon} degrees "
+                          f"for this observer within {_HORIZON_SEARCH_HOURS} hours of {utc_time}")
 
     def _scan_elevation(self, utc_time, hours, lon, lat, alt, horizon):
         """Sample the elevation above *horizon* once a minute, and find the horizon crossings.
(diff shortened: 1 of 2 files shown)


Manny7717 (approved):
Verified locally on head 2dd343e vs base 0a154c3 — implementation is correct and regression-proven.

**Regression-proven:** all 5 new tests FAIL on base (get_last_dn_time/find_aos/find_aol don't exist or are `pass` stubs) and PASS on head. Full pyorbital/tests suite: head 249 passed/1 skipped vs base 244 passed + 5 failed/1 skipped — delta is exactly the new tests, zero regressions (existing `get_last_an_time` tests still pass, so the node-search refactor preserved old numerics). ruff clean on both changed files.

**Node search (`_get_last_node_time`):** the heading-multiplication trick correctly unifies ascending/descending: for the NOAA-20 fixture, `get_last_an_time` returns the same 10:44:18.234375 as before (sub-second identical), and `get_last_dn_time` returns 09:53:29.015625 with `pos_z ≈ 0, vel_z < 0` at the crossing. Alternation probe: AN(11:00) > DN > AN(DN) > DN — nodes strictly alternate. Bisection tolerance (1 km) and 10-min walk-back step match the old code exactly.

**find_aos:** delegates to the refactored `get_next_passes` and returns `passes[0][0]`. Verified it equals the first complete-pass rise computed independently (sub-second agreement with a direct `get_next_passes(..., 24, ...)` call), and — important semantic — because `get_next_passes` only emits a pass once a set follows a rise, an in-progress pass at `utc_time` (rise already past) is naturally skipped, matching the docstring. Elevation bracketing probe: 10 s before AOS elevation = −0.565°, 10 s after = +0.574°.

**find_aol:** the scan-and-root-find path is correct — crossing indices come from `np.diff(np.sign(elev))`, `elev[crossing] > 0` selects the +→− (setting) crossings, and `_get_root`'s brentq bracket is guaranteed to straddle zero by construction. Probe: 10 s before AOL +0.592°, 10 s after −0.581°, and AOL lands inside the pass in progress at 11:00 (11:08:30) while AOS is the *next* pass (12:33:49) — exactly the documented asymmetry, and it's asserted in `test_find_aol_ends_the_pass_already_under_way`.

**Error paths:** horizon=85 (never reached for this observer/window) raises ValueError for both, message includes satellite name, horizon, and window — no silent None/IndexError. Both accept datetime, tz-aware-UTC, and np.datetime64 inputs (the `_get_tz_unaware_utctime` normalization is inherited).

**Refactor safety:** `get_next_passes`' inline scan was extracted verbatim into `_scan_elevation` (same minute sampling, same `zcs` detection, same `tol/60` root precision); a direct base-vs-head comparison of `get_next_passes` output on the fixture shows identical pass lists. Default `tol=0.001` → `_CROSSING_TOLERANCE_SECONDS` is value-identical (0.001).

Non-blocking nits (no action required):
- The 24 h window with 1-minute sampling means a pass that both rises and sets entirely between two samples is invisible — this is documented in both docstrings, just noting it also applies to very low-altitude/fast objects.
- `find_aos`/`find_aol` deliberately describe different passes when a pass is in progress (AOS = next rise, AOL = current set). Documented and tested, but callers wanting a matched pair should use `get_next_passes` directly.

Nice salvage of #206 — clean factoring, honest docstrings (the "heading" unification comment is particularly good), and tests that assert physical properties (velocity sign at node, elevation sign flip at crossing) rather than implementation details.

